PR #4, opened 2023-01-30
Simplify logic at places

piranha:
There are two commits: one about flattening `morphChildren` needs repeated `removeIdsFromConsideration` (btw, I don't cache ids and it seems to perform ok, so maybe you want to experiment with this?), and another about `syncNodeFrom` seems like a no-brainer with a little helper.

### DIFF
--- a/src/idiomorph.js
+++ b/src/idiomorph.js
@@ -269,22 +269,14 @@
                     to instanceof HTMLInputElement &&
                     from.type !== 'file') {
 
-                    let fromValue = from.value;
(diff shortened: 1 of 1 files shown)


hastebrot:
I think, this change introduced a bug when morphing an `<input>` and updating the `checked` attribute and property.

`idiomorph@0.0.8` does not have this bug, `idiomorph@0.0.9` has the bug. I need to further investigate and find the exact commit. will provide an example when I have time.

PR #12, opened 2023-08-31
Distribute library as UMD and ES packages

afcapel:
I'm interested in including idiomorph in another project that uses a JS bundler system but the current project distribution as an AMD module makes this a bit complicated.

To improve on this I've added a rollup build task that outputs different formats. At the moment it outputs a UMD and an ES modules and it drops the AMD module. AMD module popularity has declined significantly with the advent of other module systems, and providing UMD and ESM packages should cover all the cases. But let me know if that's not the case and we can add an AMD output as well.

With the new distribution files you can still use Idiomorph via a browser global by including the UMD module with a `<script>` tag, but you can also include it as an ES module and use in a project with a JS bundler, like esbuild, webpack, etc.

The htmx extension is also included as a UMD module so you can require it directly in a browser too


alexpetros:
We went through this same discussion in htmx-land, and settled just changing the AMD wrapper to UMD, which was a far, far less invasive change:

1cg:
hey @afcapel this is too violent a change for me to bring in w/o really understanding it.

i'm willing to experiment a bit w/ this repo since it's relatively small though.  i have heard good things about vite for stuff like this, what do you think?

1cg:
@afcapel we now have support for CommonJS/UMD (thank you @rzane!)  Is that "good enough" or do you want a full es module?

1cg:
hear me out:

```sh
{ echo "let Idiomorph =(" && awk "/START FACTORY/{f=1;next} /END FACTORY/{f=0} f" src/idiomorph.js && echo ")(); export {Idiomorph}" } > dist/idiomorph.esm.js
```

1cg:
OK, here's a slightly more radical change:

https://github.com/bigskysoftware/idiomorph/commit/6cc541f4fe1bd341b16003c6560d243928f00b91

Basically the idea is:

* Revert idiomorph.js to a browser-first IIFE setup
* Produce different files for each module type, AMD, UMD and ESM, by tacking on the module-type specific stuff at the end using echo

Simple and looks like it works.  Does this work for people?

afcapel:
@1cg that works for us, thanks! Love that you can get away without needing a JS bundler.

1cg:
awesome, will cut a 0.3 release now

PR #19, opened 2023-11-15
Add CommonJS support

rzane:
See also:
* https://github.com/umdjs/umd
* https://github.com/bigskysoftware/htmx/pull/1263

It seems like this PR might have a better chance of getting merged than #12, though #12 provides ESM, which would be nice to have.


1cg:
i like it, can you please make the modification only to idiomorph.js so I can merge it?

1cg:
Hi @rzane I ported over the change and have released it in v0.1.1, thank you for your contribution!

PR #33, opened 2024-01-25
Fix attribute states

dannluciano:
Without this change, after the morph, attributes (open, value, ...etc) are lost.


1cg:
this change looks really dicey to me

can you add a test showing the issue that it fixes?

dannluciano:
Yes, I'm when I understand the project test structure.

Issue #26, opened 2023-12-25
Preserving container scroll position


adicco:
I have a standard listing of audio recordings with different filters on the left which are grouped in overflowing containers; say something like this:

```html
<div id="core">
  <aside>
    <form id="author"  hx-post="...." hx-swap="morph:outerHTML" hx-target="#core" style="overflow: auto; max-height: 100px">
     <input type="checkbox" value="..." id="input-author1" /> author 1
     <input type="checkbox" value="..." id="input-author2" /> author 2
  etc...
    </div>
    
    <form id="years" hx-post="....">
    </form>
  
  </aside>

  <main>/* contents */</main>
</div>
```

Each time a checkbox is changed my backend performs a search and returns the whole #core container. This is because the filters are reactive, and indeed if you filter by a particular author, this might disable certain yeras, and vice versa (and it gets more complex with more filter categories, of course.)

I was expecting idiomorph to preserve the form's scroll position, but it doesn't. Any idea?

Thank you very much!

botandrose:
Hi @adicco, I completely agree that scroll position should be preserved! We're on the cusp of releasing v0.5.0, and the next development cycle is going to focus specifically on preserving more non-HTML-serialized state, like this.

I was aware of focus and selection preservation being an issue, but I'm surprised to learn that scroll position is also problematic! So I'm thankful for you bringing this to my attention.

PR #42, opened 2024-03-06
Fix for issue #26 to preserve overflowing containers' scroll position

KDreynolds:
I took a stab at addressing the issue user @adicco mentions in issue #26, preserving scroll position in overflowing containers.


AlexanderArvidsson:
Personally, I would probably not do this as a top-level operation, and instead as a `beforeNodeMorphed` and `afterNodeMorphed` callback.
I think this can actually be implemented completely through the callbacks without modification to the source. Not saying it shouldn't be fixed, but the approach should probably be different.

PR #69, opened 2024-12-09
Get CI green, fix config deep merging bug

botandrose:
Okay, this should get CI green again.

Two commits:
1. We have to start an http server to get the perf tests running again, because Chrome doesn't allow the fetch API over the file:// protocol. I'd like to take a look at overhauling the testing situation, but this is the smallest change I could come up with just to get us green again.
2. Bugfix: `Object.assign` is full of footguns. Passing callbacks to `Idiomorph.morph` was sometimes resulting in the default callbacks not being applied to the final config.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,10 +15,6 @@ jobs:
           cache: 'npm'
       - name: Install dependencies
         run: npm install
-      - name: Start test server
-        run: |
-          npx http-server . -p 8000 &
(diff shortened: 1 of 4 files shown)


botandrose:
Update here: I found a way to launch the Chrome instance within mocha-chrome without the security restriction for `fetch` within `file://` contexts. Much much simpler that the first strategy of booting up an http server and etc. Let me know if you want me to squash these commits.

PR #71, opened 2024-12-12
Overhaul testing framework

botandrose:
After an exhausting search through the current state of javascript testing frameworks, I settled on @web/test-runner.

Pros:
1. Run tests with Firefox and Webkit in addition to Chrome
2. Code coverage stats
3. We're able to run CI on latest node
4. The fetch permission issue with the perf tests is gone, rather than being worked around with a command line flag
5. No longer using unmaintained software (mocha-chrome)

Cons:
1. Can no longer just open the `test/index.html`  file in a browser to run the tests in a browser and debug, have to run `npm run debug` instead
2. CI takes around one minute to install dependencies, so CI runs are ~2 minutes instead of ~30 seconds.

Overall, I think its worth it. Thoughts?


botandrose:
@1cg Okay two things:
1. I've streamlined the npm test interface so now we can run e.g. `npm test test/core.js`
2. I've just pushed TESTING.md, here it is rendered: https://github.com/bigskysoftware/idiomorph/blob/71d72e2a5f345aa8c829be75f5320436f5a7b6ee/TESTING.md

How does this look now?

PR #72, opened 2024-12-12
Add `twoPass` option for additional state retention

botandrose:
# Add Two-Pass option for retaining additional element state

Superseeds #61. Read that for more context, but the gist of it is that this is a stop-gap solution for retaining more element state within morphs until `moveBefore` lands in browsers.

This draft PR is just an initial spike, so there are some missing pieces that I'd like to resolve before considering this ready for merge:
1. Only works with `morphStyle: innerHTML`
2. `moveBefore` support is still missing
3. Only tested in a super basic scenario where it successfully maintains checkbox indeterminate state

What else? Any other thoughts about this direction?

### DIFF
--- a/src/idiomorph.js
+++ b/src/idiomorph.js
@@ -70,6 +70,7 @@
  * @property {boolean} [ignoreActiveValue]
  * @property {ConfigCallbacksInternal} callbacks
  * @property {ConfigHeadInternal} head
+ * @property {boolean} [twoPass]
  */
 
 /**
@@ -78,7 +79,7 @@
  * @param {Element | Document} oldNode
  * @param {Element | Node | HTMLCollection | Node[] | string | null} newContent
  * @param {Config} [config]
- * @returns {undefined | HTMLCollection | Node[]}
+ * @returns {undefined | Node[]}
  */
 
 // base IIFE to define idiomorph
@@ -103,6 +104,7 @@ var Idiomorph = (function () {
          * @property {Set<string>} deadIds
          * @property {ConfigInternal['callbacks']} callbacks
          * @property {ConfigInternal['head']} head
+         * @property {HTMLDivElement} pantry
          */
 
         //=============================================================================
@@ -153,7 +155,7 @@ var Idiomorph = (function () {
          * @param {Element | Document} oldNode
          * @param {Element | Node | HTMLCollection | Node[] | string | null} newContent
          * @param {Config} [config]
-         * @returns {undefined | HTMLCollection | Node[]}
+         * @returns {undefined | Node[]}
          */
         function morph(oldNode, newContent, config = {}) {
 
@@ -177,7 +179,7 @@ var Idiomorph = (function () {
          * @param {Element} oldNode
          * @param {Element} normalizedNewContent
          * @param {MorphContext} ctx
-         * @returns {undefined | HTMLCollection| Node[]}
+         * @returns {undefined | Node[]}
          */
         function morphNormalizedContent(oldNode, normalizedNewContent, ctx) {
             if (ctx.head.block) {
@@ -202,6 +204,9 @@ var Idiomorph = (function () {
 
                 // innerHTML, so we are only updating the children
                 morphChildren(normalizedNewContent, oldNode, ctx);
+                if (ctx.config.twoPass) {
+                    restoreFromPantry(oldNode, ctx);
+                }
                 return Array.from(oldNode.children);
 
             } else if (ctx.morphStyle === "outerHTML" || ctx.morphStyle == null) {
@@ -220,7 +225,11 @@ var Idiomorph = (function () {
                     // if there was a best match, merge the siblings in too and return the
                     // whole bunch
                     if (morphedNode) {
-                        return insertSiblings(previousSibling, morphedNode, nextSibling);
+                        const elements = insertSiblings(previousSibling, morphedNode, nextSibling);
+                        if (ctx.config.twoPass) {
+                            restoreFromPantry(morphedNode.parentNode, ctx);
+                        }
+                        return elements
                     }
                 } else {
                     // otherwise nothing was added to the DOM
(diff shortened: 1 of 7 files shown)


src/idiomorph.js
```diff
@@ -713,11 +716,19 @@ var Idiomorph = (function () {
...
                 idMap: createIdMap(oldNode, newContent),
                 deadIds: new Set(),
+                pantry: mergedConfig.twoPass && createPantry(),
                 callbacks: mergedConfig.callbacks,
                 head: mergedConfig.head
             }
         }
 
+        function createPantry() {
+            const pantry = document.createElement("div");
```
1cg, line 726: maybe add `style='display:none'` to hide?

1cg: also maybe throw an id on it?  `idiomorph-pantry`?

src/idiomorph.js
```diff
@@ -1063,11 +1074,49 @@ var Idiomorph = (function () {
...
+            if (ctx.pantry) {
+                moveToPantry(tempNode.pantry, ctx);
+            } else {
+                tempNode.parentNode?.removeChild(tempNode);
+            }
             ctx.callbacks.afterNodeRemoved(tempNode);
         }
 
+        function moveToPantry(node, ctx) {
+            if (!node) return;
```
1cg, line 1086: rather than a recursive algo, do a query selector on `[id]` to grab all elements that have ids?

might be better for perf

src/idiomorph.js
```diff
@@ -1063,11 +1074,49 @@ var Idiomorph = (function () {
...
+                    ctx.pantry.appendChild(node);
+                }
+            }
+        }
+
+        function restoreFromPantry(root, ctx) {
+            Array.from(ctx.pantry.children).forEach(element => {
+                const matchElement = root.findElementById(element.id);
+                if (matchElement) {
+                    matchElement.before(element);
```
1cg, line 1111: looks reasonable, want to also have a `moveBefore()` branch

maybe we empty the children out of the old node when it is stuck in the pantry?

src/idiomorph.js
botandrose: Agree, except is it possible to do bottom-up in that way? Maybe that's where the idSets can come into the picture?

src/idiomorph.js
botandrose: IDK, `el.hidden = true` seems to be a higher level way to express the same thing as `style="display: none"`, and is just as good in terms of browser support  https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden#browser_compatibility

I'm not married to it, though. If you prefer the latter that's fine.

botandrose: > also maybe throw an id on it? `idiomorph-pantry`?

I actually had this at first, but ultimately decided to remove it, thinking that it could get weird with id collision if we have more than one instance of Idiomorph running on a page simultaneously, which seems plausible between Turbo and htmx. I'm not coming up with anything it buys us either, aside from someone maybe being able to pick it out visually in the dev tools html inspector while debugging. What do you think?

src/idiomorph.js
```diff
@@ -713,11 +716,19 @@ var Idiomorph = (function () {
...
+                pantry: mergedConfig.twoPass && createPantry(),
                 callbacks: mergedConfig.callbacks,
                 head: mergedConfig.head
             }
         }
 
+        function createPantry() {
+            const pantry = document.createElement("div");
+            pantry.hidden = true;
+            document.body.insertAdjacentHTML("afterend", pantry);
```
MichaelWest22, line 728: insertAdjacentElement here to fix input type here I think

src/idiomorph.js
```diff
@@ -1063,11 +1074,48 @@ var Idiomorph = (function () {
...
+            } else {
+                Array.from(node.children).forEach(child => {
+                    moveToPantry(child, ctx);
+                });
+
+                // After processing children, process the current node
+                if (node.id) {
+                    node.innerHTML = '';
+                    ctx.pantry.appendChild(node);
+                }
```
MichaelWest22, line 1102: if the current node has no id I think we have to remove it as I found otherwise non id elements remain when they shouldn't

src/idiomorph.js
```diff
@@ -1063,11 +1074,48 @@ var Idiomorph = (function () {
...
+            }
+        }
+
+        function restoreFromPantry(root, ctx) {
+            Array.from(ctx.pantry.children).forEach(element => {
+                const matchElement = root.querySelector(`#${element.id}`);
+                if (matchElement) {
+                    matchElement.before(element);
+                    element.replaceChildren(matchElement.childNodes);
```
MichaelWest22, line 1111: I found in my testing that this was adding invalid child node data but adding a ... spread operator fixed the typing for me

botandrose:
Okay, got something working with testing `moveBefore`. Ideally it'd just be a separate browser run in the main CI job, but it was surprisingly difficult to get a browser with the chrome://flags/#atomic-move experiment enabled. You can't enable it at runtime, the browser needs to boot up with that flag already enabled for it work. But there's no command-line flag to do so! The only way I could figure out how to do this was to dig into the on-disk Chrome profile and manually munge in the setting. So its currently a separate npm script: `npm run test-move-before`, and this is run as a separate job in the GitHub Actions CI.

Anyways, I've just pushed with it failing so that the next commit will correct a trivial error, so that we can see it passing, thus demonstrating that it is indeed testing what we think it is testing.

botandrose:
Just pushed a commit that makes `isSoftMatch` more strict, in that it no longer considers two elements with differing ids to be a soft match. This was a source of lost or incorrect hidden state during morphs.

botandrose:
@1cg Okay, this is ready for review! Notes since the last push:

1. I've added docs. Not my strongest suit, but they're something.
2. I added the `beforeNodePantried` hook, so Turbo can use it for nested <turbo-frame> handling.
3. I've added an env var `FORCE_TWO_PASS` to the test suite to run the entire thing with `twoPass` enabled to make sure it has the same behavior.
4. I've added two new runs to the CI matrix to test this with and without `moveBefore` added.
5. Turbo's test suite passes without changes when this version of Idiomorph is dropped in, both with twoPass on and off.

So I think we're looking good!

1cg:
💥

PR #73, opened 2024-12-17
Add tests to cover hooks, and fix extraneous hook call

botandrose:
There were no tests at all for the hooks, and hooks are about to get weird with the upcoming twoPass system, so I thought I'd add some to cover the status quo! However, this unearthed some bugs and questions.

So, three commits here:
1. First one is covering all the hook behavior, as I expected it to work, based off the README. I had to mark three tests as `skip`, because...
2. The `beforeAttributeUpdated` hook was erroneously getting called twice on additions and updates, with the second call claiming a `remove` mutation, which was definitely not happening. The change to the `src` in this PR is reflecting that fix.
3. In the last commit I change the test from what I expected the behavior to be, to what it actually is. In short, `afterNodeMorphed(oldNode, newContent)` is being called with BOTH arguments essentially being the same. On one hand, this is surprising. I'd naively expect it to show me the before and after states, especially given the argument names. However, since the first argument is the node itself, and its _after_ it's been morphed, it does make some sense for both arguments to be the same. So, is this a bug, or desired behavior? Maybe we just need to change the names of the arguments? Or just pass one `(node)` argument?

Lastly, two questions:
1. I notice that all events have a before and after hook, except for the attribute update event, which only has a `beforeAttributeUpdated` event. Should we add a corresponding `afterAttributeUpdated` event? Pushing this to its logical conclusion, it would be easy to imagine expanding this to `AttributeAdded` `AttributeUpdated` and `AttributeRemoved`, as well. Has any downstream software ever wanted any of these? 
2. Looking at the arguments for `beforeAttributeUpdated`, I expected to see the value of the attribute in there, but I do not, just the name. Should we add it?

### DIFF
--- a/src/idiomorph.js
+++ b/src/idiomorph.js
@@ -439,10 +439,10 @@ var Idiomorph = (function () {
                 // iterate backwards to avoid skipping over items when a delete occurs
                 for (let i = toAttributes.length - 1; 0 <= i; i--) {
                     const toAttribute = toAttributes[i];
-                    if (ignoreAttribute(toAttribute.name, toEl, 'remove', ctx)) {
-                        continue;
-                    }
                     if (!fromEl.hasAttribute(toAttribute.name)) {
+                        if (ignoreAttribute(toAttribute.name, toEl, 'remove', ctx)) {
+                            continue;
+                        }
                         toEl.removeAttribute(toAttribute.name);
                     }
                 }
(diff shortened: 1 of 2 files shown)


1cg:
1) no strong feelings, just wanted a way to dynamically avoid updating attributes
2) no strong feelings, I expected people to avoid updating attributes based on names, rather than values

PR #76, opened 2024-12-18
Add `--fail-only` option to fail the test suite in CI when an `it.only` is left in

botandrose:
I accidentally left an `it.only` in one of my recent commits, so CI was hamstrung without complaint. I think the presence of `it.only` should fail the CI build, to prevent this from happening again.

Two commits:
1. Create the `--fail-only` command line argument plugin, and enable it for `npm run ci`. I expect this to fail CI, because there's currently an `it.only` in `test/core.js`.
2. Remove the `it.only` and be green again.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -23,7 +23,7 @@ To run all tests against all browsers in headless mode, execute:
 ```bash
 npm run ci
 ```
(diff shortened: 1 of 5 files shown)


botandrose:
Added a commit to document the option in TESTING.md

PR #81, opened 2024-12-20
Add prettier to normalize code style

botandrose:
Okay, @1cg , here's the first sample of what a PR for this might look like. I'm betting you'll have some opinions about this!

Questions:
1. Prettier defaults to 2-space tabs, rather than your 4 spaces. Do you want this to be changed to 4 in the prettier config?
2. Any other config settings you wanna tweak?
3. Turns out it can tidy up Markdown. That seems... nice? Or should we not bother and add `*.md` it to `.prettierignore`
4. How should we enforce this, moving forward? It'd be trivial to add a job to the CI to fail the build if prettier wasn't ran, for example. But then we'd need to communicate to PR authors to be sure to run `npm run format` before committing. That's some minor added friction.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,11 @@
+node_modules/
+dist/
+coverage/
+test/*/*
+!test/lib
+package.json
+package-lock.json
+tsconfig.json
+*.md
+.github
+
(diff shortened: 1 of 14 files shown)


botandrose:
@1cg Okay, good to go. We discussed in Discord, but for posterity:
1 & 2. Sticking with defaults
3. No markdown, presentation is more important than consistency in docs
4. Not a lot of churn in this repo normally, so will be run manually before release

PR #82, opened 2024-12-23
improve softMatch to not retain state on morph if oldNode has id

MichaelWest22:
Improving softMatching so it handles not retaining state in error when morphing nodes with id's

If the old node has an id it could have some state that is important not to morph into a similar element so we can check for its ID and if either the new element has no ID or a different ID then we want to avoid morphing these nodes.  Better to instead drop the node and re-add the new content instead.

Added new test with a checkbox with indeterminate state that will not be prevented from morphing and retaining invalid state.
Also had to fix two tests that had id's on the from side of the morph which would now prevent morphing

### DIFF
--- a/src/idiomorph.js
+++ b/src/idiomorph.js
@@ -744,17 +744,21 @@ var Idiomorph = (function () {
 
         /**
          *
-         * @param {Node | null} node1
-         * @param {Node | null} node2
+         * @param {Node | null} oldNode
+         * @param {Node | null} newNode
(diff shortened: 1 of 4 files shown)


1cg:
@MichaelWest22 can you switch the order of the parameters back to the original?

1cg:
@MichaelWest22 can you update this against the latest in main?  I don't want to screw up the merge.

PR #83, opened 2024-12-23
Returning false to `beforeNodeMorphed` should affect pantry restore outcome

botandrose:
The specific motivating use-case is so that Turbo's `data-turbo-permanent` will maintain element contents as expected during `twoPass`, but it seems like the correct thing to do in general.


botandrose:
Hold up, something's not right

botandrose:
@1cg You mentioned that bugfixes didn't need your approval for merge, so I'm gonna go ahead and merge this one in.

PR #85, opened 2024-12-28
Preserve active element's focus and selection by morphing around it, when possible

botandrose:
Note: Builds on top of https://github.com/bigskysoftware/idiomorph/pull/84 , so please review just the diff of the last commit: https://github.com/bigskysoftware/idiomorph/commit/50b6e517524d854f091e2766e94b2e79862b5c8e . If #84 gets merged, I'll rebase for a easier diff.

## Overview

The new two-pass algorithm is a big improvement maintaining hidden state and element identity, but since more nodes are getting moved around vs v0.3.0, its more likely that the currently focused element (or its ancestor) will get moved, and focus and/or selection will be lost. Aside from being a regression in fidelity, this has frustrating real-world consequences, at least in the context of a real-time collaborative app.

This PR aims to preserve focus and selection across morphs by morphing around the active element and its ancestors, thus ensuring they never get moved.

## Implementation

This strategy has two steps:
1. When we build the morph context, we take the active element, find its match in the new content, and trace them both back to their respective top-most element that was passed to `Idiomorph.morph(old, new)`. If both traced paths are have the same sequence of `tagName#id`, preserving the focused element is possible, and we save a map of this information in the context.
2. Later on, while morphing, we check the map to see if we're within the critical ancestor path of the active element, and if we're about to morph the next node in that critical path. If we are, and the node is too far to the right, it would normally be moved into place, breaking focus. So if all this is true, we prime the situation by "moving" the active node into place, by repeatedly moving the current node to the end of the list until the active node is in the correct `insertionPoint` position. Once this is done, we can just continue morphing as normal. In this way, the active element and its ancestors are never moved, and focus is preserved.

## Notes / Caveats / Limitations:

When we have `moveBefore`, none of this is necessary, so we check for its existence while building the morph context, and if so none of this gets run. However, when we don't have `moveBefore`, there are three situations where this PR doesn't preserve focus. All of them might become the subject of future improvements, but they're out of scope for this PR.

1. If the active element changes levels in the DOM heirarchy, or if any element along the crticial path changes their tag name. This is simply impossible with current APIs. We could look into manually restoring focus post-morph in these cases, but again, that's out-of-scope for this PR.
2. A element along the critical path changes ids. Technically we could mutate the old id to the new id in this case and preserve focus, but that breaks our rule of maintaining identity for IDed elements. This might be reasonable to special-case, in that maybe there's no circumstance where a parent node of a focused element could have hidden state, and thus it would be safe to perform this mutation, but I haven't thought it through yet. 
3. The active element is anonymous, i.e. has no id attribute. This may or may not be solvable in a deterministic way. We might want to fall back to just trying to avoid moving the current element or its ancestors, but I haven't thought it through yet.

### DIFF
--- a/src/idiomorph.js
+++ b/src/idiomorph.js
@@ -99,6 +99,7 @@ var Idiomorph = (function () {
    * @property {ConfigInternal['morphStyle']} morphStyle
    * @property {ConfigInternal['ignoreActive']} ignoreActive
    * @property {ConfigInternal['ignoreActiveValue']} ignoreActiveValue
+   * @property {Map<Element, [Element, Element]>} activeElementMap
    * @property {Map<Node, Set<string>>} idMap
    * @property {Set<string>} persistentIds
    * @property {Set<string>} deadIds
@@ -346,22 +347,20 @@ var Idiomorph = (function () {
       oldParent = oldParent.content;
     }
 
-    /**
-     *
-     * @type {Node | null}
-     */
-    let nextNewChild = newParent.firstChild;
-    /**
-     *
-     * @type {Node | null}
-     */
-    let insertionPoint = oldParent.firstChild;
-    let newChild;
+    let insertionPoint = /** @type {Node | null} */ (oldParent.firstChild);
+    let newChild = /** @type {Node} */ ({ nextSibling: newParent.firstChild });
 
     // run through all the new content
-    while (nextNewChild) {
-      newChild = nextNewChild;
-      nextNewChild = newChild.nextSibling;
+    while ((newChild = /** @type {Node} */ (newChild.nextSibling))) {
+      // shift upcoming elements around to preserve the active element path if needed
+      if (ctx.activeElementMap.size) {
+        insertionPoint = preserveActiveElementPath(
+          oldParent,
+          insertionPoint,
+          newChild,
+          ctx,
+        );
+      }
 
       // if we are at the end of the exiting parent's children, just append
       if (insertionPoint == null) {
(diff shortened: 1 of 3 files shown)


src/idiomorph.js
```diff
@@ -425,28 +447,59 @@ var Idiomorph = (function () {
...
+   */
+  function preserveActiveElementPath(oldParent, insertionPoint, newChild, ctx) {
+    const [activeElement, newActiveElement] =
+      ctx.activeElementMap.get(oldParent) || [];
+    // are we about to morph the active element or its ancestor?
+    if (newActiveElement === newChild) {
+      while (insertionPoint && insertionPoint !== activeElement) {
+        // "move" the active element to the left by moving the current node to end of the parent
+        let nextInsertionPoint = insertionPoint.nextSibling;
+        moveBefore(oldParent, insertionPoint, null);
```
MichaelWest22, line 496:
```javascript
    if (newActiveElement === newChild) {
      const beforePoint = activeElement.nextSibling
      while (insertionPoint && insertionPoint !== activeElement) {
        // "move" the active element to the left by moving the current node after the active element
        let nextInsertionPoint = insertionPoint.nextSibling;
        moveBefore(oldParent, insertionPoint, beforePoint);
        insertionPoint = nextInsertionPoint;
      }
    }
```
if we add one line to set the beforePoint just after the activeElement then we can use this instead of null as the point to move before.  With this one line change we can minimize the movement of the nodes so they don't go all the way to the end every time

botandrose: @MichaelWest22 Love it! Pulling this in.

src/idiomorph.js
```diff
@@ -425,28 +447,61 @@ var Idiomorph = (function () {
...
+   * @param {Node} newChild
+   * @param {MorphContext} ctx
+   * @returns {Node | null}
+   * If we're about to try to morph the active element or its ancestor, indirectly "move"
+   * it into place first, so that the rest of the main loop doesn't actually move it.
+   */
+  function preserveActiveElementPath(oldParent, insertionPoint, newChild, ctx) {
+    const [activeElement, newActiveElement] =
+      ctx.activeElementMap.get(oldParent) || [];
+    if (!activeElement) return insertionPoint;
```
MichaelWest22, line 491: i think it would make more sense to move this !activeEelement check into the next if statement to become 
```javascript
if (activeElement && newActiveElement === newChild) {
```
as this check just goes straight to return insertionPoint; when it fails anyway.

PR #89, opened 2025-01-10
Restore mocha as a test runner option for easier dev debugging

botandrose:
Switching our test suite from mocha-chrome to web-test-runner has been a big win for CI:
* We're generating code coverage metrics without ceremony
* We're testing in all major browsers
* We're no longer using unmaintained software!

However, when debugging locally, web-test-runner has some tedious downsides compared to mocha:
* We have to run `npm run debug` in order to load the test
* We have to annotate the test with `it.only` if we want to run a single test 
* We have to close the browser and rerun `npm run debug` each time we make a change
* We have to remember to do a hard refresh each time or we'll get stale versions of the content

This PR restores mocha to the testing situation as an additional option, purely for use in local debugging, so we can just open `test/index.html` in a browser like before! Of course, there are some caveats:
* Perf tests are not available in mocha, due to their use of `fetch`, which is incompatible with the `file://` protocol
* There is now some duplication between the html in `test/index.html` and `web-test-runner.config.mjs`
* Mocha is not run in CI, so if the config for it breaks, we won't know until we run it manually
* We have to ensure tests remain compatible with both frameworks

I can imagine ways to address most of these caveats, but they're out of scope for this PR.

I plan to exercise my commit bit and merge this myself if there aren't any objections, so this PR is mostly a FYI for @1cg and @MichaelWest22, but please do comment if you have any feedback!

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -40,8 +40,15 @@ npm test test/core.js
 ```
 If you want to run only one specific test, you can temporarily change `it("...` to `it.only("...` in the test file, and then specify the test file as above. Don't forget to undo this before you commit!
 
-### Headed Mode (Interactive Browser Tests)
-To run tests in a non-headless mode (with a visible browser window):
+### Browser Mode
+To run tests directly in the browser, simply `open test/index.html` in a browser.
+On Ubuntu you can run:
+```bash
+xdg-open test/index.html
+```
+This runs all the tests (minus the performance tests) in the browser using Mocha instead of web-test-runner for easier debugging.
+
+If you really want to open web-test-runner in headed mode, you can run:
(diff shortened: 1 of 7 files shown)


1cg:
awesome, really like that dev workflow, makes debugging a lot easier in my experience (btw, you can run a webserver from the command line pretty easily `python3 -m http.server 8080`)

i use WebStorm which has an integrated web server so I can right click on it and open it up in a web server really efficiently

botandrose:
Awesome, I didn't know Python made it that easy! That's a super handy tool for the toolbelt.

Turns out Ruby can do it too: `ruby -run -e httpd`

PR #90, opened 2025-01-10
Use Tachometer for perf tests

botandrose:
# Overview

We have some perf tests, but they're kind of hard to interpret, due to the low signal-to-noise ratio. Thankfully, Google's Polymer team has put in the work to make a good JS performance measurement tool, [Tachometer](https://github.com/google/tachometer). This PR just does the basics for running the existing perf tests with Tachometer. Here's an example:

image

# Details
_(copied and pasted from TESTING.md)_

We have performance tests to compare Idiomorph's performance compared to morphdom. These tests are located in the `test/perf` directory. These tests are run separately from the main test suite, are not included in the coverage report, and are not run in CI. They are run with the command:
```bash
npm run perf
```

Alternatively, you can run individual performance tests with:
```bash
npm run perf checkboxes
```

You can add more tests simply by creating new `test-name.start` and `test-name.end` files in the `test/perf` directory, where the `test-name.start` file contains the starting HTML, and the `test-name.end` file contains the final HTML.

# Next Steps
This is just the groundwork. I can imagine further improvements like:
* Comparing current to previous releases, e.g. v0.3.0
* Setting up micro benchmarks to try different algorithms at the function level
* Tweaking some settings that I haven't really even looked at
* Measuring on more browsers than just Chrome

# Feedback
What do y'all think? Any drawbacks to this? Anything more you'd like to see?


botandrose:
@1cg @MichaelWest22 FYI, merging this in.

TL;DR: perf benchmarks are their own top-level thing now, separate from tests, read PERFORMANCE.md for more details.

PR #92, opened 2025-01-17
Implement better soft and id matching

MichaelWest22:
Found we can easily add back in the old matching behavior we have lost in all the refactoring quite easily. This fixes some of my tests and improves one of the focus tests as well.  Was easy to wrap it around the new core matching checking so it now handles all the edge cases I can find really well.

### DIFF
--- a/src/idiomorph.js
+++ b/src/idiomorph.js
@@ -390,6 +390,8 @@ var Idiomorph = (function () {
        */
       function findBestMatch(ctx, node, startPoint, endPoint) {
         let softMatch = null;
+        let nextSibling = node.nextSibling; 
+        let siblingSoftMatchCount = 0;
 
         let cursor = startPoint;
         while (cursor && cursor != endPoint) {
@@ -400,7 +402,7 @@ var Idiomorph = (function () {
             }
 
             // we haven't yet saved a soft match fallback
-            if (!softMatch) {
+            if (softMatch === null) {
               // the current soft match will hard match something else in the future, leave it
               if (!ctx.idMap.has(cursor)) {
                 // optimization: if node can't id set match, we can just return the soft match immediately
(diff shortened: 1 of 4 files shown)


botandrose:
Hi @MichaelWest22, thanks for taking the time to explore this! However, as I mentioned in Discord, I believe these bailout mechanisms were designed to deal with a trade-off that only existed in v0.3.0, and no longer exists in the current algorithm. I'll outline my understanding here, and if you see any flaws in it, please do point them out!

In the v0.3.0 world, using a match that was far ahead in the childNode list meant **literally deleting** any nodes in-between the insertion point and the match, and then they would be recreated from scratch later, minus any hidden state. So there was an inherent tension between a) selecting the best match, and b) avoiding deleting other potentially important nodes. So the code in this PR was designed precisely to try to strike a balance within this tension, by being careful to not look ahead too far, and being mindful of what was going to be deleted.

However, the new algorithm no longer deletes nodes in-between the insertion point and the match, it just **moves them out of the way**. If the node is persistent, we move it to the pantry. If the node will be a soft-match, we just bump it to the end of the children so it can be soft-matched later. With this strategy, that previous trade-off is gone, and there is no longer any significant downside to always choosing the best match. IMO, this is the primary win of the new algorithm, in terms of correctness, and it seems to me that restoring this code would undermine that correctness while also adding complexity.

All that said, looking at the scenario you have in the soft match test on this branch, I think it does demonstrate that there is some room for improving the soft-match selection, particularly in the pathological case where a new anonymous node is shifted to the front of a set of anonymous sibling nodes that share its nodeType. I'd be happy to see a PR designed to address this, but only after this long-running branch gets merged, as it seems to me to be a matter of optimization, not a matter of correctness.

So in the meantime, I'm going close this PR. But if you think I've missed something in my reasoning above, please do reopen!

PR #94, opened 2025-01-30
improve id matching during findBestMatch and fix hook case

MichaelWest22:


MichaelWest22:
https://codepen.io/MichaelWest22/pen/myboqmO

Here is a code pen example of the textbook example from the Readme of video preservation you can do with idiomoprh.  When you swap the heading div between above and below the video it should allow the id in the video to be respected and the heading div should not be picked as a good match to morph with a lower div that would destroy the id'ed videos state.  With the current algorithm it just always finds the first match for every new node no matter how destructive it would be to id'ed or other nodes.  By adding back the original idset matching logic that was really the core of what idiomorph was designed for it works perfectly for all situations the old idiomoprh routine could retain state like videos.

https://codepen.io/MichaelWest22/pen/RNbdjJp

Here is the same idiomoprh textbook example with this pr code showing it works great moving the header up and down even when using normal browsers without moveBefore feature just like it always did.

Also there is a focus test that was broken that is now fixed you can check out in this PR where now focus is preserved correctly even when the browser does not have moveBefore.  

There are also going to be many situations where without this change in place preservation hooks that used to work to preserve nodes will just not longer work properly.  When id's or nodes are reordered or moved it will discard a lot of the nodes in the middle and the preservation hooks will retain the element but it will then leave it in the wrong dom location.  We should be trying to keep hooks and preservation the same as the old algorithm where possible so we dont break any existing apps and use cases.  We may have better id node state preservation but for some new situations but keeping old correctness is also important.

Issue #27, opened 2023-12-28
preserve input value if no attr change


hesxenon:
Since it is possible(-ish) to work around this feel free to close this issue, but imho the current behaviour around input values is wrong/confusing.

Consider the following structure: 

```html
<form name="form">
      <input id="input" placeholder="Test">
</form>
```

If I type something into this input and then do

```js
Idiomorph.morph(document.form, document.form.outerHTML)
```

I'd expect the inputs value to be preserved since nothing changed and an inputs `value` is only loosely coupled to the attribute of the same name, similar to the play state of a video.

According to [this part](https://github.com/bigskysoftware/idiomorph/blob/61327049948378b4f7ba0d4c29419420986d8a79/src/idiomorph.js#L362) the value of an input is reset if 
- the new content (`from`) does not have a `value` attribute
- the `beforeAttributeUpdate` callback does not return false

Imho it would be better to simplify this to a mismatch in the value attribute like this:

```js
if(from.getAttribute("value") !== to.getAttribute("value") && !beforeAttributeUpdate(attr, to, 'remove')) {
  to.value = from.getAttribute("value") ?? ""
}
```

meaning that the behaviour can be depicted as the following table
| value attribute of `to` (existing) | value attribute of `from` (new content) | behaviour |
| --- | --- | --- |
| null (not present) | null (not present) | no desire to change -> preserve state |
| some value | the same value | no desire to change -> preserve state |
| some value | null (not present) | desire to reset -> set to '' |
| null (not present) | some value | desire to change -> set to value of `from` |
| some value | some other value | desire to change -> set to value of `from` |

hesxenon:
I've forked your repository and after some experiments my manual tests seem to do what I'd expect. You can find the source here: https://github.com/hesxenon/idiomorph/blob/main/src/idiomorph.js#L350

I'd be glad to open a PR if you agree with the premise of this issue _and_ if you tell me how to run your test suite :wink: - because I only got a conn refused from mocha-chrome :sweat_smile:

1cg:
hey @hesxenon I htink this should at least be an option.  we are in the process of reworking the test suite, are you OK with revisiting it when we are done?  Probably early 2025 timeframe?

hesxenon:
sure

svefro:
+1 for this. if old element does not have a value and new element does not have a value, user input should not be reset.

svefro:
[...]

PR #95, opened 2025-01-31
Ignore value feature

MichaelWest22:
Here is a possible new feature we can add.  We already have support to ignore the value of the active element.  This existing feature allow you to morph say a form with inputs to a new state but it will not update the contents of the item the user is currently focused on and entering input into.  However for all other inputs in the same form their current values will need to passed back to the server and returned as they are in the response that gets morphed in so that the state of the other inputs can be preserved.  This probably works well for most cases.  But it would also be nice to support a more generic value preserving option like this IgnoreValue option where instead it can preserve ALL inputs, checkboxes and textarea's with the current DOM entry state.  I found this feature mainly useful when testing form replacement and mutations on the client side as I didn't have a server interaction to submit and recreate the input values. I do not know if there is wider demand for this feature outside of this so feel free to ignore and close this PR


MichaelWest22:
Found this issue #27 that has a similar suggestion and need for this same feature.  There is a suggestion in this issue to re-write the sync input code and remove and simplify most of the logic to handle situations where inputs have been changed by end users but not the attributes themselves.  But this is likely a breaking change for some users so would have to be added as an option.  But my solution here also seems to cover their use case well and I think is less likely to risk causing issues as my change allows you to override all input value swapping if this is the requirement instead of tring to do a smarter optional attribute comparision.

Issue #40, opened 2024-02-22
morph:<expr> not usable when using hx-swap-oob with htmx WebSocket extension


aral:
The issue with `hx-swap` properties not being available in `hx-swap-oob` (https://github.com/bigskysoftware/htmx/issues/2308) also affects idiomorph.

Specifically, the following syntax cannot be used:

```html
hx-swap-oob='morph:<expr>'
```

Use case:  I was just trying to make a markdown preview updated via WebSocket scroll into view to show the latest change after the content has been morphed and I can’t.

aral:
So the fix for this that I implemented in my fork of htmx-2 (which has the hx-swap-oob normalisation PR merged in so it differs from current htmx-2 a bit), was essentially to special-case idiomorph:

```js
(!swapStyle.startsWith('morph:') && swapStyle.indexOf(':') > 0)
```

at the equivalent of the following line:

https://github.com/bigskysoftware/htmx/blob/bf692737011b40316397d55c3ee756fde6c502cf/src/htmx.js#L1417

This leaves the selector alone and allows it to be parsed properly by idiomorph.

PR #97, opened 2025-01-31
support full extended swap styles in htmx oob swaps

MichaelWest22:
Currently oob swaps in htmx are not supported with idiomorph if you want to do a inner style morph as they include a : which is a seperator in hx-swap-oob. So to support this i've allowed : to be replaced with ;; in the swap style so it can also be used for supporting extended config inside the swap style as well.

For example here is a now supported extended swap with hx-swap-oob:
```html
<div hx-swap-oob="morph;;{morphStyle;;'innerHTML'}:#div">
```


MichaelWest22:
should resolve #40 hopefully

botandrose:
Hmm... this looks to me like something that should be fixed in htmx, rather than worked-around in Idiomorph.

@1cg what do you think about resolving it via https://github.com/bigskysoftware/htmx/issues/2308 ?

botandrose:
@MichaelWest22 I'm closing this in favor of fixing the issue in htmx.

PR #101, opened 2025-02-07
Bugfix: insertionPoint can become incorrect after morphNode and createNode

botandrose:
Redux of https://github.com/bigskysoftware/idiomorph/pull/99

This is a very subtle bug! Let's say we perform the following morph, wrapping an IDed node in a `<p>`:
`<div><a id="a"></div>` => `<div><p><a id="a"></p></div>`

When the `<p>` is created, `insertionPoint` points to `#a`. We then descend into the new `<p>` and use `moveBeforeById` to pull the `#a` into it, preserving its identity. However, when we then ascend back out into the parent `<div>` and continue the morph, `insertionPoint` still points to `#a`, which is all wrong. This can also happen in a different branch in a more convoluted scenario. 

To address this issue, we simply reset the `insertionPoint` after each mutating operation.

Big props to @MichaelWest22 for discovering this and coming up with a solution.

### DIFF
--- a/src/idiomorph.js
+++ b/src/idiomorph.js
@@ -323,11 +323,21 @@ var Idiomorph = (function () {
             ctx,
           );
           morphNode(movedChild, newChild, ctx);
+          insertionPoint = movedChild.nextSibling;
           continue;
         }
 
         // last resort: insert the new node from scratch
-        createNode(oldParent, newChild, insertionPoint, ctx);
+        const insertedNode = createNode(
+          oldParent,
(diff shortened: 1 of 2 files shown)


botandrose:
@MichaelWest22 as I mentioned in Discord, I wanted to redo the specific bugfix implementation a bit, for more CQRS in our functions. I also reworded some of the descriptions a bit. Do you see anything I missed or was otherwise lost from your original PR?

PR #102, opened 2025-02-10
Fix outerHTML morph when newContent has siblings

botandrose:
While testing the upcoming v0.5.0 release, @delaneyj found that an incorrect morph can occur when all of the following is true:
1. outerHTML morphStyle
2. newContent is type Node
3. newContent has a parentNode
4. newContent has sibling nodes that aren't part of the morph

Turns out this is due to us reusing the `newContent`'s `parentNode` for outerHTML calculations, and that is not safe to do in this specific case, because `newContent`'s siblings can leak into the morph. I haven't found a slam-dunk solution to this, but I've come up with the following strategies, all with their own drawbacks:
1. Serialize `newContent` to an HTML string, and then reparse that and use it as the normalized newContent. Since `newContent` is attached to the DOM in this case, there could be hidden state that we want to persist, so this would lose all that. Not to mention the added cost of serialization and deserialization.
2. Clone the newContent tree and reparent it to a document-fragment. More or less the same drawbacks as above.
3. Restore the v0.4.0 outerHTML morph code. I think this is a last resort, in that its a whole separate bespoke algorithm just for this one outerHTML case, and I'd rather have one single well-tested codepath for all morphing.

Instead, the solution I'm going with here is to create a duck-typed fake parent node that responds with the correct values. It has none of the drawbacks above, but its kind of a hack, and we gotta tell the typechecker to look the other way. Hopefully we can come up with something better in a future refactoring.

### DIFF
--- a/src/idiomorph.js
+++ b/src/idiomorph.js
@@ -1129,7 +1129,15 @@ var Idiomorph = (function () {
       /** @type {Map<Node, Set<string>>} */
       let idMap = new Map();
       populateIdMapForNode(oldContent, oldElts, persistentIds, idMap);
-      populateIdMapForNode(newContent, newElts, persistentIds, idMap);
+
+      let newRoot = newContent;
+      // if newContent is a duck-typed parent, pass its single child node as the root to halt upwards iteration
+      /** @ts-ignore */
+      if (newContent.__idiomorphDummyParent) {
+        newRoot = /** @type {Element} */ (newContent.childNodes[0]);
+      }
+      populateIdMapForNode(newRoot, newElts, persistentIds, idMap);
+
       return { persistentIds, idMap };
     }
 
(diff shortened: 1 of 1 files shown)


MichaelWest22:
I had a quick stab at that test and investigated how it was handled in older releases.  I found in old releases it was not perfect as it does actually mutate the newContent passed in by appending the node to a dummy parent which can remove it from its source.  I don't think this was ever a deal breaker as the new source content has always been disposable.

Then I found from my testing that during outerHTML morphing we now normalize the parent of the old content which we are unable to re-parent without breaking things so the code in normalizeParent now just returns the parentNode for the Node case.  However in all cases I found that the oldNode input is required to always be an Element and it will always have a valid parentNode which can be an Element or a Document or a DocumentFragment.  Its always safe to just use parentNode and cast it as a Element because all the actions we perform on it as a ParentNode.  In fact the correct JSDoc type should actually be ParentNode and not Element for many of the internal fields we use but fixing this was a much bigger change and pretending everything is Element is just cleaner.  

https://github.com/bigskysoftware/idiomorph/compare/main...MichaelWest22:idiomorph:fix-outer-normalize

Here is my example of the simple fix I found.  Just fixing the normalization of oldNode and reverting the change made to normalizeParent Node handling so it will always append the node into a div.  Updated the test to show what happens here

Then I tested how we could fix the mutation of the newContent if we really needed this as an upgrade over old versions and I came up with:

https://github.com/bigskysoftware/idiomorph/compare/main...MichaelWest22:idiomorph:fix-outer-normalize2

this alternative change where we can clone the node to solve that case.  Also we can check if the node is a single child and deal with the happy path of returning the parent.

botandrose:
@MichaelWest22 Amazing work, thank you! As usual, you've spotted nuances I've missed. :)

Since I'm trying to get v0.5.0 out the door, and thus trying to make the most conservative fixes possible, I'm just going to fix the issue in `querySelectorAll` that you spotted, and then merge this branch in as the next release candidate for v0.5.0.

However, after we release, I want to investigate ditching this code for your solution in  `idiomorph:fix-outer-normalize`. I don't want to merge it for v0.5.0, because it does break some potential use-cases, like morphing within the same document with hidden state. I don't know if anyone is actually doing this (and I hope they aren't), but I don't want to introduce another regression and block release further. After all, this bug itself is an example of someone doing something I never thought would happen!

Regarding `fix-outer-normalize2`, I really like it from a functional point of view, but from a performance standpoint, I'm concerned that cloning the entire tree would increase GC pressure, and I'm starting to suspect we might already have a big problem with that in general. More investigation needed there for sure. If it turns out not to be expensive, it seems like a great solution.

I've already merged the newElts/oldElts fix into main, since that seems like a no-brainer.

Finally, I'm going to take a look your `fix-idMap` branch separately.

PR #118, opened 2025-02-17
retain focus by breaking on active during findBestMatch

MichaelWest22:
I found that by getting the findBestMatch to break out of its searching loop when it finds the active element it will then either match on the current softmatch or return null which then causes any required elements to be inserted before the active element instead of displacing the active focused element.  With this we are able to retain full focus and selection in almost all cases possible.  Focus is still lost without moveBefore or restoreFocus in situations where we can't retain focus like moving an input to different levels in the DOM. Also it does still lose focus in the tests where the focused input is moved down to another div but a preservable id element is left in the old div which still matches first causing possible focus loss.

I Updated and renamed all the many now fixed tests.  Note this does leave us with some redundancy in our tests we could probably clean up later.

### DIFF
--- a/src/idiomorph.js
+++ b/src/idiomorph.js
@@ -432,6 +432,10 @@ var Idiomorph = (function () {
             }
           }
 
+          // if the current node contains active element, stop looking for better future matches,
+          // because if one is found, this node will be moved to the pantry, reparenting it and thus losing focus
+          if (cursor.contains(document.activeElement)) break;
+
           cursor = cursor.nextSibling;
         }
 
--- a/test/preserve-focus.js
+++ b/test/preserve-focus.js
@@ -22,7 +22,7 @@ describe("Preserves focus algorithmically where possible", function () {
     }
   }
 
-  it("preserves focus state and outerHTML morphStyle", function () {
+  it("preserves focus state when elements are swapped", function () {
     assertFocusPreservation(
       `
       <div>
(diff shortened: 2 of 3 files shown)


botandrose:
@MichaelWest22 Looks really good so far, but I notice an `it.only` was left in. It turns out that I broke the CI detection of this on `main`, but I've just pushed a new commit fixing it. So now, can you please remove the `it.only`, rebase the branch on current `main`, and then force push it?

MichaelWest22:
aww crap.  i tried to run test:ci which would have caught this but it fails because prettier checks apply to ALL files which breaks and seems to ignore the prettier ignore file for some reason for me

botandrose:
Huh, yeah, that's annoying. Feel free to swap the order of the checks in the `test:ci` script in package.json, if its better for you.

MichaelWest22:
No actually it seems prettier ignore is working fine it is just swapping files between CRLF and LF mode annoyingly causing my issues

botandrose:
Huh, wow, it turns out I can force-push _your_ branch. Probably because `Maintainers are allowed to edit this pull request` is checked? Anyways, I squashed it all together into one commit. I won't do this in the future without asking! I'm honestly surprised it let me

botandrose:
I've played around with it, and it seems like a clear improvement. Very clever! Ultimately, its just preventing pantrying (i.e. reparenting) of the active element, is that correct?

botandrose:
This seems like a clear win, and its so simple! Really great job.

My only hesitation is that I have a vague memory of moving from the future in the middle of iterating childnodes causing correctness issue. I think it would yield incorrect html when `beforeNodeRemoved => false` was used. I need to get a test up for that anyways, because it impacts another one of your PRs, and even one of my own for that matter.

Let me get that test up, and if it doesn't break with this PR, it'll be an easy merge.

botandrose:
You know what, I think your #99 PR fixed the issue I was remembering, too.

Merging this! Great work, Michael.

PR #120, opened 2025-02-18
only restore selection when needed in restoreFocuse

MichaelWest22:
Found that when restoring focus selection we also need to check that the element doesn't already have a selection set before we update it because it is possible the user could have change the selection during the morph. so just checking the current selectionEnd is not valid before restoring the selection seems to fix the issue.

 Also found a bug where if the selection begins at the 0 position the if statement was treating 0 as false for selectionStart which prevents selection restoration.  We only need to verify the selectionEnd value is not 0 or undefined so just removed the selectionStart check.

Added a simulated test for updating selection during morph that was failing before the fix and also changed one of the other tests to have a selection starting at position 0.

### DIFF
--- a/src/idiomorph.js
+++ b/src/idiomorph.js
@@ -216,8 +216,10 @@ var Idiomorph = (function () {
    */
   function saveAndRestoreFocus(ctx, fn) {
     if (!ctx.config.restoreFocus) return fn();
-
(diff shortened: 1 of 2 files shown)


botandrose:
@MichaelWest22 Great work, once again!

PR #121, opened 2025-02-20
Restore direct imports, and add named export for ESM htmx extension

### LINKED ISSUE #38: htmx extension without globals
The htmx init code here relies on a `htmx` global: https://github.com/bigskysoftware/idiomorph/blob/8e40c42cc573609eb6863e72fa3403574974dd7d/src/idiomorph-htmx.js#L12

While reliance on globals is probably okay with `<script>` usage, in a scenario with a bundler it is not necessary and [undesired](https://stackoverflow.com/questions/10525582/why-are-global-variables-considered-bad-practice) to have any globals. How about allowing something like this?

````js
import htmx from 'htmx.org';
import idiomorph from 'idiomorph/dist/idiomorph-ext-bundler.min.js';
htmx.defineExtension('morph', idiomorph);
````

Related: https://github.com/bigskysoftware/htmx-extensions/issues/128

### LINKED ISSUE #119: Trouble getting Idiomorph Setup w/ Django + Webpack
I'm trying to get idiomorph to do full page swaps on the body -- using django + webpack

My body tag is as follows:

```html
<body class="bg-base-100" hx-headers='{"X-CSRFToken": "{{ csrf_token }}"}' hx-ext="morph" hx-swap="morph">
```

My htmx.js file which gets included via webpack:

```js
import htmx from "htmx.org"
import 'idiomorph';
import { Idiomorph } from 'idiomorph';

window.htmx = htmx
window.Idiomorph = Idiomorph;
```

My form with all the inputs:

```html
<form hx-get="/" hx-target="body" hx-trigger="change target:input, change target:select, submit" hx-push-url="true" class="mb-4" id="search-form">
    // many input fields 
</form>
```

And a demo of this setup failing to use idiomorph:

https://github.com/user-attachments/assets/5598c95d-cccf-4aa1-a59f-8daf225f95ff

I'm truly unsure of what I could be missing. Idiomorph exists in the browser console, so I'm assuming it's failing to register via the hx- tags on the body? Not sure if I need to also import the idiomorph-htmx.js or the idiomorph-ext.js?

botandrose:
1. Fixes a regression in v0.7 where one could no longer import dist paths, e.g. `import "idiomorph/dist/idiomorph.esm.js"`
2. Adds an official export path for Idiomorph + htmx extension: `import "idiomorph/htmx"`

Closes #119
Closes #38

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,12 @@ the `dist/idiomorph-ext.js` file in your HTML:
 </div>
 ```
 
+or by importing the "idiomorph/htmx" module:
+
+```html
+import "idiomorph/htmx";
+```
+
 Note that this file includes both Idiomorph and the htmx extension.
 
 #### Configuring Morphing Behavior in htmx
--- a/package.json
+++ b/package.json
@@ -19,8 +19,12 @@
   "module": "dist/idiomorph.esm.js",
   "unpkg": "dist/idiomorph.min.js",
   "exports": {
-    "require": "./dist/idiomorph.cjs.js",
-    "import": "./dist/idiomorph.esm.js"
+    ".": {
+      "require": "./dist/idiomorph.cjs.js",
+      "import": "./dist/idiomorph.esm.js"
+    },
+    "./htmx": "./dist/idiomorph-ext.esm.js",
+    "./dist/*": "./dist/*"
   },
   "scripts": {
     "test": "web-test-runner",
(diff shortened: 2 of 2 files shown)


botandrose:
@MichaelWest22 Can you take a look at this branch with your local webpack and vite setup, and verify I didn't screw it up? Note that you'll need to run `npm run dist` to generate the ESM file targeted by the `idiomorph/htmx` export path.

MichaelWest22 (commented):
Yeah did a quick test in webpack and vite and it seems to be working fine

PR #129, opened 2025-03-03
Rework duck-typed parent into SlicedParentNode class

### LINKED ISSUE #124: `hx-*` attributes ignored when swapping in different type of element
When using `hx-swap="morph"` to swap in a different type of element (e.g. a `<div>` to replace a `<form>`), the `hx-*` attributes of the new element are ignored and the element becomes inert.

The following examples demonstrate the problem:
- [A minimal reproduction of the problem using 0.7.2](https://codepen.io/timomihaljov-yle/pen/WbNGqpb?editors=1000)
- [The same code working as expected with 0.4.0](https://codepen.io/timomihaljov-yle/pen/raNMXev?editors=1000)

botandrose:
The main difference here is that `childNodes` is now a live calculation, based on a bounding box defined by the originally-passed node's siblings. In other words, a parent node containing a live slice of its children.

Also, the `SlicedParentNode` name is better and more intention-revealing, I think.

Lastly, it is now a JavaScript `class`, which definitely goes against the overall C-ish style of Idiomorph. But since we're emulating a concrete aspect of the OOP DOM API, I think it makes sense.   

Fixes #124, and the return value issue in #126.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Fixed:
   * Fix error when morphing elements with numeric ids (@botandrose, @ksbrooksjr)
   * Fix issue with outerHTML morphing an IDed node that gets moved (@botandrose, @MichaelWest22)
+  * Fix incorrect return value when root element gets moved or replaced in an outerHTML morph (@botandrose, @MichaelWest22)
 
 ## [0.7.2] - 2025-02-20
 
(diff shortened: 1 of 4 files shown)


botandrose:
`npm run perf v0.7.2` shows no detectable perf difference from v0.7.2 on any benchmark, which is what I think we should expect.

Its slower to recalculate the `childNodes` each call, as opposed to returning a static array, of course. But that recalculation should only happen a few times per morph when we're morphing outerHTML style, and thus be negligible.

All benchmarks ARE using the `outerHTML` morph style, though, so this code is definitely getting run!

Issue #130, opened 2025-03-11
Using the .id getter in the isSoftMatch can be problematic when another js frameworks override it


dombesz:
### The Problem

I'm using angular frontend components that are being sent to the frontend via turbo.js, morphed with Idiomorph.
The result of the morphing will insert duplicate elements in case of angular components. This happens because, the `isSoftMatch` function uses the `id` as comparison as `(!oldElt.id || oldElt.id === newElt.id)` and angular can override the `id` getter.

In my case, the angular components coming via turbo do not have an html `id` attribute, but they have a `@Input id = "default-id"` property in the component definition instead.. Because of this, angular will dynamically assign an `id` to the component, and calling `oldElt.id` will return `"default-id"`, although no html id attribute was provided in the component template. 

This is problematic because the `newElt` component is yet to be inserted to the DOM and angular did not initialise it component yet. As a result, calling `newElt.id` will not be `"default-id"`, but it will be the id provided as the html attribute on the response, which is an empty string in my case.

### Proposed solution

In order to avoid the interference of initialized vs not initialized angular components, I propose to compare the `id` html attributes directly by calling `oldElt.getAttribute("id")` instead of `oldElt.id` in the [`isSoftMatch`](https://github.com/bigskysoftware/idiomorph/blob/a6ada993e8c8d7cb30f417983d3b003982d2e9d0/src/idiomorph.js#L466-L479) method:

```javascript
 function isSoftMatch(oldNode, newNode) {
  // ok to cast: if one is not element, `id` and `tagName` will be undefined and we'll just compare that.
  const oldElt = /** @type {Element} */ (oldNode);
  const newElt = /** @type {Element} */ (newNode);

  return (
    oldElt.nodeType === newElt.nodeType &&
    oldElt.tagName === newElt.tagName &&
    // If oldElt has an `id` with possible state and it doesn't match newElt.id then avoid morphing.
    // We'll still match an anonymous node with an IDed newElt, though, because if it got this far,
    // its not persistent, and new nodes can't have any hidden state.
    (
      (!oldElt.getAttribute("id")) ||
      (oldElt.getAttribute("id") === newElt.getAttribute("id"))
    )
  );
}
```

I can provide a PR if this turns out to be a confirmed bug.
Thanks!

botandrose:
Should be resolved by #136

botandrose:
@dombesz https://www.npmjs.com/package/idiomorph/v/0.7.4

PR #131, opened 2025-03-11
Use getAttribute when soft matching ids.

dombesz:
Some javascript frameworks like angular can overwrite the tag's id getter, and comparison can turn into a false negative. Newly added elements don't have the framework's code initialised yet, and calling id on those elements can differ from the id assigned via angular. Since there is no way to have frameworks initialised on the new elements, calling the `id` can be inconsistent. Using getAttribute("id") ensures that we always compare the html id attributes.

See this issue for more details: #130


botandrose:
Hey @dombesz , thanks for the PR! A couple of notes before I review:

1. Don't modify the files in `dist/`, we rebuild these only just before a release. 
2. Can you get CI green? You can do a simulation locally with `npm test:ci`. More details in TESTING.md

botandrose:
@dombesz Hello! Thank you for your time and effort here, and sorry for the long wait on this!

The problem you pointed out turned out to be more widespread than javascript frameworks like Angular... its an issue with the DOM API itself! Take a look at #135 and #136 for more details. I expect #136 will resolve this as well, do you agree? Can you test the branch in your app to verify?

I'm going to run this branch in production on my app for the next week, and if all looks well, I'll push a bugfix release next sunday.

botandrose:
closing in favor of #136

dombesz:
Hi @botandrose , yes I believe the issue is solved by your PR as well. Thanks!

botandrose:
@dombesz great! I'm glad to hear that.

FYI, I've found another unrelated bug that I want to get into the next bugfix release, so contrary to what I wrote last Sunday, I expect to cut a bugfix release NEXT Sunday now.

PR #136, opened 2025-07-20
Fix persistence bug caused by untrustworthy HTMLFormElement.id getter

### LINKED ISSUE #135: Form with id field
Here is a kind of fun bug.
line 477:
```javascript
/ If oldElt has an `id` with possible state and it doesn't match newElt.id then avoid morphing.
// We'll still match an anonymous node with an IDed newElt, though, because if it got this far,
// its not persistent, and new nodes can't have any hidden state.
(!oldElt.id || oldElt.id === newElt.id)
```

When oldElt or newElt is a form element with an input field that has `name="id"`, then `elt.id` will return the input field.
This should probably be `!oldElt.getAttribute("id") || oldElt.t.getAttribute("id") === newEltt.getAttribute("id")`

Edit: Clarified that it's `elt.id` not `elt.input`.

botandrose:
Turns out we can't simply call `.id` on element nodes, because that property is unreliable on `<form>` and `<fieldset>` elements. It could return the id of the node as expected, OR (problematically) a child element that has a name attribute of `id`. The latter breaks persistence of the form element, which can have further consequences involving lost state.

This PR has two commits:
1. A failing test exposing the issue
2. A commit that fixes it, but at the cost of a 5% perf hit on my machine.

I'm marking this PR as WIP in the hopes I can find a faster way to resolve this.

For more information, see: https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement#issues_with_naming_elements
Fixes #135

### DIFF
--- a/src/idiomorph.js
+++ b/src/idiomorph.js
@@ -225,7 +225,7 @@ var Idiomorph = (function () {
 
     const results = fn();
 
-    if (activeElementId && activeElementId !== document.activeElement?.id) {
+    if (activeElementId && activeElementId !== id(document.activeElement)) {
       activeElement = ctx.target.querySelector(`[id="${activeElementId}"]`);
       activeElement?.focus();
     }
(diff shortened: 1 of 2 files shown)


botandrose:
Interestingly, it appears that the majority of the perf loss is simply in the function call itself... if I replace the implementation of the new `id` function with simply `return node.id`, the perf loss is about the same! I expected that JIT inlining would negate this, but apparently not.

Maybe a manual inline then? I don't love it, since we're accesing element ids all over the place... hmm.

botandrose:
Okay, I got something working here with inlining the function. Perf is mostly unaffected now, and the bug is fixed, but code readability suffers. Hmm.

PR #149, opened 2026-05-05
Skip beforeAttributeUpdated when an attribute value is unchanged

monorkin:
This is a small optimization that can significantly reduces morph time on dense pages without changing semantics to consumers.

`morphAttributes` calls the `beforeAttributeUpdated` callback for every attribute on every element being morphed, before checking whether the new value differs from the existing one. This PR changes it to trigger the callback only when the new and old values differ.

From what I read in abf6c94, the callback's purpose is to allow consumers to ignore changes to attributes - when the values are the same there is nothing to prevent.

### DIFF
--- a/src/idiomorph.js
+++ b/src/idiomorph.js
@@ -676,12 +676,13 @@ var Idiomorph = (function () {
         const oldAttributes = oldElt.attributes;
         const newAttributes = newElt.attributes;
         for (const newAttribute of newAttributes) {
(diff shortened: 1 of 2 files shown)


botandrose:
@monorkin Hello Stanko, great find! This looks like a bug, honestly... the `beforeAttributeUpdated` hook should only fire if the attribute is going to be updated.

What's more, I've been hoping to find ways to reduce the total cost of the callback hooks for other reasons, so I'm personally excited for raising the performance ceiling of Idiomorph in this way.

Great work!

PR #150, opened 2026-05-23
Fix TypeError in saveAndRestoreFocus when restored id resolves to a non-input element

emaia:
## Problem

`saveAndRestoreFocus` can throw `TypeError: activeElement.setSelectionRange is not a function` during a morph.

After the morph, the function re-queries the previously-focused element by its saved id:

  ```js
  activeElement = ctx.target.querySelector(`[id="${activeElementId}"]`);
```

It then calls `activeElement.setSelectionRange(...)`. Two problems compound here:

1. querySelector returns Element | null — any element sharing that id matches  (div, span, button, …). The instanceof `HTMLInputElement | HTMLTextAreaElement`  guard at the top of the function is only applied to the pre-morph element and is never  re-applied after the re-query.

2. `setSelectionRange` is called with no support check. The existing !activeElement.selectionEnd guard does not protect against non-inputs, since selectionEnd is undefined on them (making !undefined truthy).

Additionally, setSelectionRange throws InvalidStateError on real inputs whose type doesn't support text selection (number, email, date, …).

This is reachable from any morph where a focused input's id no longer maps to an input-like element in the new DOM. It surfaced in the wild via Turbo's morph refresh method (<turbo-stream action="refresh">), where auto-generated input ids are common. The morph still completes, but it produces a noisy unhandled promise rejection.

## Fix

- Re-apply the input/textarea instanceof check after the re-query; fall back to null otherwise.
- Feature-detect setSelectionRange and wrap the call in try/catch to absorb the InvalidStateError thrown by non-selectable input types.

## Tests

Added two cases to test/restore-focus.js:

- saved id resolves to a `div` after morph — reproduces the exact TypeError on main and passes with the fix.
- focused type=number input — guards the unsupported-type path.

### DIFF
--- a/src/idiomorph.js
+++ b/src/idiomorph.js
@@ -230,11 +230,27 @@ var Idiomorph = (function () {
       activeElementId &&
       activeElementId !== document.activeElement?.getAttribute("id")
     ) {
-      activeElement = ctx.target.querySelector(`[id="${activeElementId}"]`);
+      const candidate = ctx.target.querySelector(`[id="${activeElementId}"]`);
+      // the re-queried element may not be an input/textarea even if the
+      // original was, since any element can share the saved id
+      activeElement =
+        candidate instanceof HTMLInputElement ||
+        candidate instanceof HTMLTextAreaElement
+          ? candidate
+          : null;
       activeElement?.focus();
     }
-    if (activeElement && !activeElement.selectionEnd && selectionEnd) {
-      activeElement.setSelectionRange(selectionStart, selectionEnd);
+    if (
+      activeElement &&
+      typeof activeElement.setSelectionRange === "function" &&
+      !activeElement.selectionEnd &&
+      selectionEnd
+    ) {
+      try {
+        activeElement.setSelectionRange(selectionStart, selectionEnd);
+      } catch {
+        // setSelectionRange throws on input types that don't support text selection
+      }
     }
 
     return results;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by botandrose** [comment]
@emaia Hi Emaia, thank you for this PR! It looks good to me on first glance. Allow me to restate the issue to make sure I understand correctly.

This PR address two different, but closely-related bugs:
1. `saveAndRestoreFocus` throws trying to focus a unfocusable element, when the morph changed the active element from a focusable element to an unfocusable element
2. `saveAndRestoreFocus` throws trying to set a selection on an unselectable element, if the morph changed the active element from a selectable element to an unselectable element

The overall behavior is still correct, but there are errors in the console, and this PR results in those errors being silenced.

Is this is all correct?

**Comment by botandrose** [comment]
@emaia Looking into this a bit more closely, I'm not sure we need to protect against `focus()` being called at all, even if the element changes. Calling `div.focus()` for example throws no error. Or am I missing something?

**Comment by botandrose** [comment]
@emaia I also wonder if we can simplify the `setSelectionRange` guards. It looks like there are two levels of checks:

Simplifying:
```js
if(typeof activeElement.setSelectionRange === "function") {
  try {
    activeElement.setSelectionRange(selectionStart, selectionEnd);
  } catch {}
}
```
It seems like the `try ... catch` check makes the `type === "function"` check redundant. Can we get away with just the try catch?

emaia:
@botandrose Good catch! Yes, we can drop it. I've simplified to just the try/catch. Thanks!

botandrose:
@emaia Merged. Thank you very much!

Issue #147, opened 2025-12-05
setAttribute error when morphing elements with Alpine.js shorthand syntax (@change, :class, etc.)


Gloib:
When using Idiomorph with Alpine.js, morphing fails when server-returned content contains Alpine's shorthand attribute syntax (e.g., @change) on non-input elements (like div, span, etc.).

Error:
`InvalidCharacterError: Failed to execute 'setAttribute' on 'Element': '@change' is not a valid attribute name.`

```
<!-- Original -->
<div>Content</div>

<!-- Server returns morphed content with shorthand syntax -->
<div @change="doSomething()">Content</div>
```

The issue is that setAttribute() validates attribute names strictly, while the browser's HTML parser may have accepted invalid names when parsing the server response.

For now I'm using the full Alpine syntax `x-on:change` instead of shorthands `@change` in server responses. It would be nice to support shorthands since they work in the initial DOM and are idiomatic Alpine.js.

<img width="971" height="202" alt="Image" src="https://github.com/user-attachments/assets/f587c993-0e51-4a22-97ab-d89f453aac27" />

botandrose:
Hi @Gloib Thanks for bringing this to our attention! I see the value in supporting the shorthand attributes, so if we can support them in a reasonable fashion, then I'd like to. The question then becomes, is this even possible?

Let's say we need to reuse an existing node to maintain its hidden state, perhaps a `<video>` element that is currently playing. If the morph were to want to add a `@change` attribute to this element, is it even possible to do so via some other mechanism than `setAttribute`? Maybe something like:

```js
for (const a of Array.from(source.attributes)) {
  target.setAttributeNode(a.cloneNode());
}
```

That might work?

Failing that, maybe one could create some kind of plugin for Idiomorph and Alpine.js that interprets and rewrites the shorthand versions into their long html-compliant form?

What are your thoughts? I'm pretty busy here at the moment with Real Life stuff, and I have some big plans for the next version of Idiomorph that I want to tackle first. So if you have the time to experiment with finding a solution here that'd really help get this out the door sooner.

botandrose:
I've just merged more-or-less what I described above into `main`. Some interesting notes here:
1. Chromium relaxed their attribute parsing, so this has already "fixed itself" on that browser. But Firefox and Safari still benefit from switching to `setAttributeNode`, so its still worth doing.
[...]

PR #156, opened 2026-08-13
Recognize nodes from other realms via realm-safe type checks

lizarusi:
Hi!

While using idiomorph at Walnut, we hit a bug that we've been carrying as a local patch — contributing the fix upstream.

## The bug

Passing a node created in another JS realm (e.g. an iframe's `contentDocument`) as `newContent` throws:

```
TypeError: newContent is not iterable
    at normalizeParent (src/idiomorph.js)
    at Object.morph (src/idiomorph.js)
```

Repro:

```js
const target = document.querySelector("#target");
const iframe = document.createElement("iframe");
document.body.append(iframe);
const foreign = iframe.contentDocument.createElement("button");
foreign.textContent = "Bar";
Idiomorph.morph(target, foreign); // 💥 TypeError: newContent is not iterable
```

Every realm has its own constructors, so cross-realm nodes fail `instanceof Node` in `normalizeParent` and fall through to the array/HTMLCollection branch, which blows up on the spread.

The crash is only the visible part. `instanceof` stays false for these nodes **even after they are adopted into this document**, so once past normalization they also silently fail every other realm-sensitive `instanceof` check: template content handling, id-based matching of new children (`newChild instanceof Element`), and `syncInputValue` — meaning input/option/textarea values would silently not sync from cross-realm content. The second test demonstrates that.

We hit this in production at Walnut, where we morph DOM captured from customer apps across document boundaries — we've been carrying a fix as a local patch since 0.7.2.

## The fix

Replace realm-sensitive `instanceof` checks with tiny duck-typing helpers (`nodeType` / `localName`) — the same approach morphdom uses. The helpers carry JSDoc type predicates, so TypeScript narrows exactly like `instanceof` did and `npm run typecheck` stays green. The `document.activeElement` checks in the focus-preservation code keep `instanceof`, since the active element always belongs to this document.

First commit adds the failing tests, second commit makes them pass. There is a dedicated test per converted check — entry normalization, id-based element preservation (both the dummy-parent and `SlicedParentNode` paths), template content morphing, input/option/textarea value syncing, head handling, and Document normalization — and each test was verified to fail if its single check alone is reverted to `instanceof`. `npm run typecheck`, `npm run format:check`, and the full suite pass, with coverage at 100%.

Relates to #103 — whichever direction the `newContent` type-narrowing goes, cross-realm nodes currently crash rather than being handled or rejected cleanly, so this seems worth fixing today.

### DIFF
--- a/src/idiomorph.js
+++ b/src/idiomorph.js
@@ -136,15 +136,15 @@ var Idiomorph = (function () {
   //=============================================================================
 
   /**
-   * @param {any} value
+   * @param {unknown} value
    * @returns {value is Node}
    */
   function isNode(value) {
-    return typeof value?.nodeType === "number";
+    return typeof (/** @type {any} */ (value)?.nodeType) === "number";
   }
 
   /**
-   * @param {any} value
+   * @param {Node | null | undefined} value
    * @returns {value is Element}
    */
   function isElement(value) {
@@ -152,51 +152,66 @@ var Idiomorph = (function () {
   }
 
   /**
-   * @param {any} value
+   * @param {Node | null | undefined} value
    * @returns {value is Document}
    */
   function isDocument(value) {
     return value?.nodeType === Node.DOCUMENT_NODE;
   }
 
+  const HTML_NAMESPACE = "http://www.w3.org/1999/xhtml";
+
   /**
-   * @param {any} value
+   * @param {Node | null | undefined} value
+   * @param {string} localName
+   * @returns {value is Element}
+   */
+  function isHtmlElement(value, localName) {
+    return (
+      isElement(value) &&
+      value.localName === localName &&
+      value.namespaceURI === HTML_NAMESPACE
+    );
+  }
+
+  /**
+   * @param {Node | null | undefined} value
    * @returns {value is HTMLTemplateElement}
    */
   function isTemplateElement(value) {
-    return isElement(value) && value.localName === "template";
+    return isHtmlElement(value, "template");
   }
 
   /**
-   * @param {any} value
+   * @param {Node | null | undefined} value
    * @returns {value is HTMLHeadElement}
    */
   function isHeadElement(value) {
-    return isElement(value) && value.localName === "head";
+    return isHtmlElement(value, "head");
   }
 
   /**
-   * @param {any} value
+   * @param {Node | null | undefined} value
    * @returns {value is HTMLInputElement}
    */
   function isInputElement(value) {
-    return isElement(value) && value.localName === "input";
+    return isHtmlElement(value, "input");
   }
 
   /**
-   * @param {any} value
+   * @param {Node | null | undefined} value
    * @returns {value is HTMLOptionElement}
    */
   function isOptionElement(value) {
-    return isElement(value) && value.localName === "option";
+    return isHtmlElement(value, "option");
   }
 
   /**
-   * @param {any} value
+   * @param {Node | null | undefined} value
    * @returns {value is HTMLTextAreaElement}
    */
   function isTextAreaElement(value) {
-    return isElement(value) && value.localName === "textarea";
+    return isHtmlElement(value, "textarea");
   }
   /**
    * Default configuration values, updatable by users now
--- a/src/idiomorph.js
+++ b/src/idiomorph.js
@@ -124,6 +124,80 @@ var Idiomorph = (function () {
   //=============================================================================
 
   const noOp = () => {};
+
+  //=============================================================================
+  // Realm-safe node type checks.
+  //
+  // `instanceof Node` (and friends) fails for nodes created in another JS
+  // realm (e.g. an iframe's document), even after they are adopted into this
+  // document, because each realm has its own constructors. These helpers
+  // duck-type nodes via `nodeType` and `localName` instead, so nodes from any
+  // realm are recognized.
+  //=============================================================================
+
+  /**
+   * @param {any} value
+   * @returns {value is Node}
+   */
+  function isNode(value) {
+    return typeof value?.nodeType === "number";
+  }
+
+  /**
+   * @param {any} value
+   * @returns {value is Element}
+   */
+  function isElement(value) {
+    return value?.nodeType === Node.ELEMENT_NODE;
+  }
+
+  /**
+   * @param {any} value
+   * @returns {value is Document}
+   */
+  function isDocument(value) {
+    return value?.nodeType === Node.DOCUMENT_NODE;
+  }
+
+  /**
+   * @param {any} value
+   * @returns {value is HTMLTemplateElement}
+   */
+  function isTemplateElement(value) {
+    return isElement(value) && value.localName === "template";
+  }
+
+  /**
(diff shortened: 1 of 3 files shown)


src/idiomorph.js
```diff
@@ -120,6 +120,80 @@ var Idiomorph = (function () {
...
+  // realm are recognized.
+  //=============================================================================
+
+  /**
+   * @param {any} value
+   * @returns {value is Node}
+   */
+  function isNode(value) {
+    return typeof value?.nodeType === "number";
+  }
```
myabc, line 147:
```suggestion
  /**
   * @param {unknown} value
   * @returns {value is Node}
   */
  function isNode(value) {
    return typeof (/** @type {any} */ (value)?.nodeType) === "number";
  }
```

src/idiomorph.js
```diff
@@ -120,6 +120,80 @@ var Idiomorph = (function () {
...
+
+  /**
+   * @param {any} value
```
myabc:
```suggestion
   * @param {Node | null | undefined} value
```

src/idiomorph.js
```diff
@@ -120,6 +120,80 @@ var Idiomorph = (function () {
...
+   * @returns {value is Element}
+   */
+  function isElement(value) {
+    return value?.nodeType === Node.ELEMENT_NODE;
+  }
+
+  /**
+   * @param {any} value
```
myabc:
```suggestion
   * @param {Node | null | undefined} value
```

src/idiomorph.js
```diff
@@ -120,6 +120,80 @@ var Idiomorph = (function () {
...
+    return isElement(value) && value.localName === "option";
+  }
+
+  /**
+   * @param {any} value
+   * @returns {value is HTMLTextAreaElement}
+   */
+  function isTextAreaElement(value) {
+    return isElement(value) && value.localName === "textarea";
+  }
```
myabc, line 218:
```suggestion
  const HTML_NAMESPACE = "http://www.w3.org/1999/xhtml";

  /**
   * @param {Node | null | undefined} value
   * @param {string} localName
   * @returns {value is Element}
   */
  function isHtmlElement(value, localName) {
    return (
      isElement(value) &&
      value.localName === localName &&
      value.namespaceURI === HTML_NAMESPACE
    );
  }

  /**
   * @param {Node | null | undefined} value
   * @returns {value is HTMLTemplateElement}
   */
  function isTemplateElement(value) {
    return isHtmlElement(value, "template");
  }

  /**
   * @param {Node | null | undefined} value
   * @returns {value is HTMLHeadElement}
   */
  function isHeadElement(value) {
    return isHtmlElement(value, "head");
  }

  /**
   * @param {Node | null | undefined} value
   * @returns {value is HTMLInputElement}
   */
  function isInputElement(value) {
    return isHtmlElement(value, "input");
  }

  /**
   * @param {Node | null | undefined} value
   * @returns {value is HTMLOptionElement}
   */
  function isOptionElement(value) {
    return isHtmlElement(value, "option");
  }

  /**
   * @param {Node | null | undefined} value
   * @returns {value is HTMLTextAreaElement}
   */
  function isTextAreaElement(value) {
    return isHtmlElement(value, "textarea");
  }

```

myabc: 1. While `localName === "template"` might more performant than `instanceof HTMLTemplateElement`[^1], that check by itself will match non-HTML namespaces.

    Since SVG/namespace landed in #147, #154, this could be problematic. To reproduce:

    ```js
    const SVG = "http://www.w3.org/2000/svg";
    let oldT = document.createElementNS(SVG, "template");
    oldT.appendChild(document.createElementNS(SVG, "g"));
    document.body.appendChild(oldT);
    let newT = document.createElementNS(SVG, "template");
    newT.appendChild(document.createElementNS(SVG, "rect"));
    Idiomorph.morph(oldT, newT);
    ```

    Thus `isTemplateElement` returns `true` for an `SVGElement`, `morphChildren` does `oldParent = oldParent.content` leading to `undefined` and a `TypeError` (I think?)
    
    My suggestion would be to add a namespace guard.

2. I'd replace `any` type for predicate params with `Node | null | undefined` value as above (except for `isNode` ofc).

[^1]: conjecture - I haven't actually measured!

myabc:
@lizarusi I'm not a maintainer - just a contributor/community member - so please take my drive-by suggestions with a grain of salt! That said - with a few fixes to the JSDocs and a guard for HTML docs - this patch has my vote!

myabc:
Additionally, `restoreFocus` hardcodes `document.activeElement` (`src/idiomorph.js:301`) rather than `ctx.target.ownerDocument`. This would need to be fixed to fully support morphing foreign documents.

lizarusi:
@myabc Thanks a lot for looking into this! I've applied your suggestions (the `isHtmlElement` namespace guard and the JSDoc types) and added tests to verify them, plus fixed `restoreFocus` to resolve the active element via `ctx.target.ownerDocument` as you pointed out.

Additionally, I decided to keep `value instanceof <constructor>` as a first check before duck-typing: property probes can be fooled by form named-property shadowing (e.g. a form containing `<input name="nodeType">`), which `instanceof` is immune to — duck-typing now only kicks in for nodes from other realms.

src/idiomorph.js
lizarusi: amazing findings! all suggested fixes applied

src/idiomorph.js
```diff
@@ -207,18 +299,15 @@ var Idiomorph = (function () {
    */
   function saveAndRestoreFocus(ctx, fn) {
     if (!ctx.config.restoreFocus) return fn();
+    // focus lives in the target's document, which may not be ours
+    const doc = ctx.target.ownerDocument;
```
myabc: 👍🏻

botandrose:
@lizarusi @myabc Thanks for the effort here, folks! I agree that supporting cross-realm morphs is something Idiomorph should seamlessly support. It took me a little while to review this, because I had a few concerns:
1. Does adding these checks regress performance?
My tests indicate no.
2. Does the morph leave the morphed tree in a state of mixed-realm nodes?
This seems like it would be a really hard-to-detect footgun for other totally unrelated code doing `instanceof` checks. Looks like morphing from another realm to the main document is no problem, but the reverse does have this issue. We'll want to address that, if possible, or warn and advise, if not. But let's do that in a follow up PR.
3. Does this add complexity to the codebase, or make it less readable?
It is more lines, but I actually think the readability is improved, at the cost of some indirection. I am going to refactor a bit to not blow out the main IIFE namespace with a bunch of utility functions, and move the block lower, so that the start of the file is still the main `morph` function.
4. Is this a real issue hit by real people? Or an AI doing a drive-by PR for a bug that technically exists, but isn't actually affecting anyone?
I do see that AI was used, but aside from the PR text being obviously generated, it was readable enough. And ultimately it looks like you are owning the work yourself, presenting it to me as yourself, being responsive in the comments, doing the rebases, etc. Mentioning that you were hitting it yourself in a real-world app helps, too.

So all-in-all, great work!

PR #159, opened 2026-08-26
Track generated `dist/*.d.ts` alongside the other artifacts

myabc:
> [!NOTE]
> Missed as part of #153.

Running the build leaves `dist/idiomorph.d.ts` and `dist/idiomorph-ext.d.ts` permanently untracked, so `git status` is never clean afterwards:

```
$ git status --short
?? dist/idiomorph-ext.d.ts
?? dist/idiomorph.d.ts
```

They are the only thing `npm run dist` produces that was never added to git. Nothing ignores them either, so they just sit there as noise. This adds both files.

Tracking them because they are generated from `dist/*.esm.js` rather than from `src/` means they move in lockstep with the artifacts already committed. They add no churn outside the release commit, which already touches all of `dist/`. It also keeps types working for anyone installing straight from the repository, since `package.json` points `"types"` at `dist/idiomorph.d.ts`.

The alternative was `dist/*.d.ts` in `.gitignore`. I checked that would not break publishing because the `files` allow-list wins over `.gitignore`, so `npm pack` still includes both. This splits `dist/` into some artifacts tracked and some ignored. I will switch if you would rather not carry the files.

### DIFF
--- a/dist/idiomorph-ext.d.ts
+++ b/dist/idiomorph-ext.d.ts
@@ -1,0 +1,156 @@
+export type ConfigHead = {
+    style?: "merge" | "append" | "morph" | "none";
+    block?: boolean;
+    ignore?: boolean;
+    shouldPreserve?: (arg0: Element) => boolean;
+    shouldReAppend?: (arg0: Element) => boolean;
+    shouldRemove?: (arg0: Element) => boolean;
+    afterHeadMorphed?: (arg0: Element, arg1: {
+        added: Node[];
+        kept: Element[];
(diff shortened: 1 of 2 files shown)


myabc:
@botandrose In line with [your comment on build artifacts](https://github.com/bigskysoftware/idiomorph/pull/153#issuecomment-5307501604), I didn't end up committing any files in `dist/`, but I think it probably does make sense to go ahead and track `dist/*.d.ts` until a release comes out (the churn should be minimal).

PR #160, opened 2026-08-26
Declare package as ESM so `import` works on older Node

### LINKED ISSUE #158: ESM entry is not declared as a module, so `import { Idiomorph }` fails on Node without syntax detection
### Describe the bug

The `import` condition resolves to `dist/idiomorph.esm.js`, but the package has no `"type": "module"` and the file has a `.js` extension. Per Node's rules that makes it an *ambiguous* file, which Node treats as CommonJS unless [module syntax detection](https://nodejs.org/en/blog/release/v22.7.0) rescues it. On any Node without syntax detection, `import { Idiomorph } from "idiomorph"` fails to link:

```
SyntaxError: Named export 'Idiomorph' not found. The requested module 'idiomorph' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'idiomorph';
const { Idiomorph } = pkg;

    at ModuleJob._instantiate (node:internal/modules/esm/module_job:146:21)
```

Syntax detection was unflagged in **22.7.0** and backported to **20.19.0**, so this affects Node 18.x, 20.0–20.18, and 22.0–22.6. I hit it in a deployed serverless runtime while it worked on local Node 22 — the classic "works on my machine" split, since the two Nodes disagree about the format of the very same file.

Worth noting this is *not* the same as #91 (bundler/Vite resolution) — this is Node's own ESM loader, and it happens with no bundler involved.

### Steps To Reproduce

```sh
mkdir idiomorph-repro && cd idiomorph-repro
npm init -y && npm pkg set type=module
npm install idiomorph@0.7.4
printf 'import { Idiomorph } from "idiomorph";\nconsole.log(typeof Idiomorph.morph);\n' > index.js

npx -y node@20.18.0 index.js   # SyntaxError: Named export 'Idiomorph' not found
npx -y node@20.19.0 index.js   # function
```

Same split on a single modern Node, since the flag toggles exactly the mechanism that papers over this:

```sh
node index.js                                   # function
node --no-experimental-detect-module index.js   # SyntaxError
```

### Expected Behavior

`import { Idiomorph } from "idiomorph"` links on every supported Node, without relying on syntax detection to guess the format.

### Suggested fix

Declare the format explicitly, either:

1. add `"type": "module"` to `package.json` — `dist/idiomorph.js` and the `<script>`/`unpkg` builds are loaded by extension or by browsers, so they aren't affected by the field; or
2. rename the ESM entry to `dist/idiomorph.mjs` and point the `import` condition at it.

Node [explicitly recommends](https://nodejs.org/en/blog/release/v22.7.0) the former anyway, since detection costs an extra parse: *"add `\"type\": \"module\"` to the nearest parent package.json file to eliminate the performance cost."*

This looks worth doing before 0.8.0 ships. 825388e removes the `require` condition and `main` for #122, which leaves the mis-typed ESM file as the *only* entry point — so on the affected Node versions there'd no longer be a CommonJS path to fall back to. Adding `"type": "module"` in the same release would make that removal safe.

myabc:
Fixes #158.

`exports["."].import` resolves to `dist/idiomorph.esm.js`, but the package has no `"type"` field and the file has a `.js` extension. Node calls that an *ambiguous* file and treats it as CommonJS unless [module syntax detection](https://nodejs.org/en/blog/release/v22.7.0) rescues it. Detection was unflagged in 22.7.0 and backported to 20.19.0, so on Node 18.x, 20.0-20.18 and 22.0-22.6 `import { Idiomorph } from "idiomorph"` fails to link with `SyntaxError: Named export 'Idiomorph' not found`.

A repro with no flags and no bundler:

```sh
npm init -y && npm pkg set type=module
npm install idiomorph
echo 'import { Idiomorph } from "idiomorph"' > index.js

npx -y node@20.18.0 index.js   # SyntaxError: Named export 'Idiomorph' not found
npx -y node@20.19.0 index.js   # works
```

Same package, same file. The two Nodes disagree about its format because the package doesn't declare one. 825388e changed this: dropping the `require` condition and `main` leaves the ESM file as the only entry point, so those versions have no CommonJS path left to fall back to.

## The fix

`"type": "module"` in `package.json` is what Node recommends over relying on detection, which costs an extra parse. It also disambiguates `dist/idiomorph.js` and `dist/idiomorph-ext.js`. There are no dist renames, and no change to `files`, `module`, `unpkg`, `types` or `exports`.

The field applies to every `.js` file in the package, so the two in-tree CommonJS scripts had to go. Rather than rename them to `.cjs`, they're converted to ESM: `perf/runner.js` resolves its two paths against `import.meta.url` instead of `__dirname`, and `test/lib/ensure-full-coverage.js` takes `lcov-parse`'s CommonJS export as a default import. Both are dev-only and ship in neither the published tarball nor the browser tests, which are served as `<script>` tags and never see `package.json`. `<script src>` and unpkg consumers are unaffected because browsers pick format by MIME type rather than by `package.json`.


botandrose:
@myabc Thank you for explaining this, and for the PR.

On one hand, those node versions are EOL, but on the other hand, its so simple to add `"type": "module"` that we might as well just do it, if there aren't any drawbacks.

Seems like the only drawback is needing to rename those two remaining CJS files to .cjs so that they still work. And really, I want idiomorph to only use ESM and <script src> so really I think the fix there would be to convert those files to ESM, thus finally closing the door on CJS. Is that reasonable to do here?

I would prefer to not have the extra github action. So drop that and I'll merge, and if you feel like taking a crack at the CJS removal, go for it. Either here in this PR, or in a follow up, or I'll just merge the CJS rename and do it myself.

myabc:
@botandrose thanks for the feedback!

> I want idiomorph to only use ESM and <script src> so really I think the fix there would be to convert those files to ESM, thus finally closing the door on CJS. Is that reasonable to do here?

tbh I prefer that approach too. I just wanted to keep this PR tightly-scoped.

I'll drop the GitHub Action. I think we can trust that ESM imports "just work" on all recent node versions once `"type": "module"` is in place.

botandrose:
@myabc Looks great! Thank you!

PR #163, opened 2026-09-01
Read `ownerDocument` off the Node prototype

myabc:
> [!NOTE]
> This and #164 are pre-emptive fixes to possible issues that I found when reviewing @lizarusi's #156 

A form control named `ownerDocument` shadows the form's DOM property, causing morphing to fail when creating the pantry in the target document.

Resolve `ctx.doc` through the native `Node` getter. Cache it on first use so importing the module does not require a DOM.

Tests cover focus and selection restoration on a shadowing form and loading without DOM constructors.

### DIFF
--- a/src/idiomorph.js
+++ b/src/idiomorph.js
@@ -1459,13 +1459,13 @@ var Idiomorph = (function () {
    * instead, which is also realm-safe: it brand-checks the internal slot
    * rather than the realm the node was created in.
    */
-  /** @type {(() => Document) | undefined} */
+  /** @type {((this: Node) => Document) | undefined} */
   let ownerDocumentGetter;
 
   /** @param {Node} node @returns {Document} */
   const ownerDocumentOf = (node) => {
     // Resolve lazily so importing the module does not require a DOM.
-    ownerDocumentGetter ??= /** @type {() => Document} */ (
+    ownerDocumentGetter ??= /** @type {(this: Node) => Document} */ (
       /** @type {PropertyDescriptor} */ (
         Object.getOwnPropertyDescriptor(Node.prototype, "ownerDocument")
       ).get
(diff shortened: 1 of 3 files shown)


myabc:
@botandrose @lizarusi apologies that these patches didn't make the cut for v0.8.0 (although tbf I don't think they are critical). I had Claude also take a look at #156, but sat on the findings until I could verify.

Would be grateful for your feedback!

botandrose:
@myabc Hi Alex, thank you for exploring this one!

While this is indeed a bug, its such an extreme corner-case, that I think the juice on this one is just not worth the squeeze. It seems to me that someone has to essentially _try_ to hit this, and its not easy to hit! They have to have a very unlikely setup with a form element named `ownerDocument`, be morphing the form specifically. And then even if this happens, the actual breakage is simply the focus restoration, which itself is only needed for non-`moveBefore` browsers like Safari, and only as a backup if the nature of the morph happens to re-parent the focused element.

So I would bet this bug has probably literally never happened in the real world, and the consequences are minor anyways.

Furthermore, the upcoming Safari 26.6 / 27.0 is supposed to support `moveBefore`, at which point I'm going to rip out this entire focus-save-and-restore mechanism anyways, because it'll no longer be needed.

So I'm going to close this one. Let me know if you disagree!

myabc:
@botandrose I understand the trade-off between covering all edge cases vs. keeping a solution clean and easy to reason about, so 👍🏻 from me for closing these two PRs.

While we have hit occasional issues with form DOM properties colliding method names at @opf, this particular edge case hasn't caused us issues.

> Furthermore, the upcoming Safari 26.6 / 27.0 is supposed to support moveBefore, at which point I'm going to rip out this entire focus-save-and-restore mechanism anyways, because it'll no longer be needed.

Nice!

lizarusi:
@myabc thanks for looking into this! Yeah, very interesting findings, I agree that even though these are real bugs, you'd almost have to be trying to break it to hit these cases. 

Let's see if anyone hits them in the wild!

PR #164, opened 2026-09-01
Harden ID lookup against form control shadowing

myabc:
> [!NOTE]
> This and #163 are pre-emptive fixes to possible issues that I found when reviewing @lizarusi's #156 

> [!NOTE]
> Depends on #163; its changes are included until that PR merges.

Read ID attributes and query ID-bearing elements through native methods to avoid form controls shadowing `getAttribute` or `querySelectorAll`. Resolve those methods on first use to preserve imports without a DOM. Restore the comment explaining the realm-safe type checks.

Tests cover form method shadowing and extend the realm suite with value-attribute, textarea, and algorithmic focus assertions. Upstream now supplies the target-document focus implementation. Broader form attribute-method shadowing remains outside this change.

### DIFF
--- a/src/idiomorph.js
+++ b/src/idiomorph.js
@@ -1459,13 +1459,13 @@ var Idiomorph = (function () {
    * instead, which is also realm-safe: it brand-checks the internal slot
    * rather than the realm the node was created in.
    */
-  /** @type {(() => Document) | undefined} */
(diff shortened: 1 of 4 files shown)


botandrose:
@myabc Hi Alex, thanks again for your effort here. I've to the same conclusion here as in #163 . Let me know if I've missed something, or if you disagree.

